Issue #13, opened 2017-10-25
Slow Performance when switching sketches


karismatic-megafauna:
### Context
I have a simple App that switches between sketches on click:

code: https://github.com/michaelghinrichs/i-hope-this-art
example: https://michaelghinrichs.github.io/i-hope-this-art/ 

### Problem
I have started to notice if i switch between sketches (click the screen 10 times thus changing the current sketch 10 times) the canvas starts to get really really slow... try it out on the site and notice that the drawings render very slowly in a direct correlation to how many times you have switched sketches.

### What I am doing wrong?
Currently, i just pass a different function to `sketch`, is that a bad practice? Should each sketch be it's own p5 wrapper?

Here is what my code looks like:
```jsx
import React, { Component } from 'react';
import P5Wrapper from 'react-p5-wrapper';
import doubleCircles from './drawings/doubleCircles';
import contours from './drawings/contours';
import flower from './drawings/flower';
import './App.css';

 class App extends Component {
   constructor(props) {
     super(props);
     this.state = {
       sketches: [ contours, flower, doubleCircles ],
       active: flower,
     }
   }

   switchSketch = () => {
     const current = this.state.active;
     const nextSketchs = this.state.sketches.filter((sketch) => {
       return sketch.name !== current.name;
     });
     const newSketchs = nextSketchs.concat(current);
     this.setState({
       active: nextSketchs[0],
       sketches: newSketchs,
     });
   }

   render() {
     return (
       <div className="App" onClick={this.switchSketch}>
         <P5Wrapper sketch={this.state.active} />
       </div>
     );
   }
 }
```

Any idea what I am doing wrong? Should I unmount and remount the `P5Wrapper`?

au-re:
Encountered the same problem, created a pull request that should solve this: #15

and-who:
I just released a new Version since Ages 😅 
Please check if your Issue got fixed.
If not please reopen this Issue or create a new one.

PR #15, opened 2017-12-22
fix: #13 performance degradation

au-re:
Removing the dom node does not remove the JS objects associated to it causing a memory leak (see eugene lazutkin comment https://stackoverflow.com/questions/683366/remove-all-the-children-dom-elements-in-div). 

Using `this.canvas.remove()` instead of `this.wrapper.removeChild(this.wrapper.childNodes[0]);` seems to fix this.


and-who:
Sry, that this took Ages! 😅 
I just released a new Version, so this should be fixed.

PR #19, opened 2018-01-24
easy cleanup when component unmounts

slin12:
Simplifies the cleanup needed when the P5 Wrapper unmounts and prevents memory leaks. Instead of having to deal with cleanup in the sketch, will automatically remove the canvas in the wrapper.


and-who:
Sry, that this took Ages! 😅 
I just released a new Version, so this should be fixed.

PR #65, opened 2020-11-25
Refactor for v3

### LINKED ISSUE #12: Pass function as property?
Can I also pass a function as a property let's say something like this:

```
<P5Wrapper sketch={sketch} action={actionFunction} />

//sketch
export function sketch(P) {
  const p = P;

  p.setup = function (props) {
    p.createCanvas(width, height);
   props.actionFunction();
  };

  p.draw = function () {
    }
  };
}
```

I've tried and failed. Is it possible to do this?

### LINKED ISSUE #72: Type signatures for TypeScript projects
Any thoughts on publishing type signatures for this project? I've worked around this locally by doing...

`lib/types.ts`
```
import * as p5 from 'p5';
import { ComponentType } from 'react';

export type P5WrapperComponent = ComponentType<{ sketch: (p: p5) => void }>;
```

`index.d.ts`
```
declare module 'react-p5-wrapper' {
  import { P5WrapperComponent } from './lib/types';
  export default P5WrapperComponent;
}
```

### LINKED ISSUE #73: Bundle process size and performance issues
Currently the bundler in use is webpack. The end build is quite sizeable, relatively speaking, and the build time can be a couple of seconds.

Areas of potential improvement:

1. Code splitting
2. Async loading of dependencies
3. New tools to replace current ones

Improving this process and adding a watcher to the dev process would make the developer experience when managing the library much better.

jamesrweb:
The basic idea is to move towards a more functional approach. As you can see the code is a lot tighter than before but still retains its functionality. There is one major issue still outstanding and that is the rerendering issue. I need to look into this but you can run the demo and see that it does apply the rotation as expected albeit with a rerender for each frame (which we don't want). Will keep working on this over the coming while but appreciate any feedback on the pros/cons of this implementation and also any ideas on how to resolve the rerendering issue that you may have. Until then I will see what I can do on my end and update as I go.


jamesrweb:
Released the main changes as part of #101. A few more changes will be added before the version 3 release but we are almost there!

PR #69, opened 2020-12-02
Allow custom attributes on the wrapper

### LINKED ISSUE #67: ClassName prop
It would be nice to be able to pass a className prop to style the wrapper div directly.

jamesrweb:
Adds the ability for users to pass an  props to the wrapper instance and each of these attributes will be in turn put onto the wrapper such as id / class / etc.


jamesrweb:
Fixes #67.

PR #79, opened 2021-03-28
Add linting and formatting

yevdyko:
This MR adds linting and formatting to simplify the maintenance of the library component.

#### What was changed?

 - Add Prettier and its config for formatting.
 - Format component using Prettier.
 - Add pre-commit git hook that formats staged files.
 - Add ESLint packages for typescript as a linter.
 - Add ESLint config.
 - Add React specific linting rules for ESLint.
 - Add packages to integrate ESLint with Prettier rules.
 - Add CI Github action workflow.
 - Add documentation about ESLint and Prettier configurations.
 - Format the whole project using prettier.
 - Remove unused styles.
 - Fix linting for example js files.


.eslintignore
```diff
@@ -0,0 +1,3 @@
+example/dist
+dist
```
jamesrweb, line 1: I would like it if we could have the example compile to dist when published to gh-pages and then the actual code published to dist when we run the publish script, do you think you could add this to this PR or should we move this to a different one? I am open to both options.

.eslintrc
```diff
@@ -0,0 +1,28 @@
+{
```
jamesrweb, line 1: Can you add documentation for the configuration into `docs/tools/eslint.md`?

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
```
jamesrweb, line 1: Can you add documentation for the configuration into `docs/tools/prettier.md`?

package.json
```diff
@@ -74,7 +83,19 @@
...
     "deploy": "gh-pages -d example/dist",
     "publish-demo": "npm run build-demo && npm run deploy",
-    "publish-npm-package": "npm run build && npm publish"
+    "publish-npm-package": "npm run build && npm publish",
+    "format": "prettier --config .prettierrc --parser typescript --write 'src/**/*.tsx'",
+    "lint": "eslint 'src/**/*.tsx' --ext .ts",
+    "lint:fix": "eslint 'src/**/*.tsx' --ext .ts --fix"
+  },
+  "simple-git-hooks": {
+    "pre-commit": "npx lint-staged"
```
jamesrweb: I would prefer that formatting, testing, etc is done in CI and not with git hooks. Mostly because it reduces dependencies and offsets formatting, testing and building to the integration suite.

Could you create a configuration for [GitHub Actions](https://github.com/features/actions)? I have [an example workflow file](https://github.com/jamesrweb/codewars-api-client/blob/master/.github/workflows/CI.yml) here from another side project of mine for reference.

src/index.tsx
```diff
@@ -25,8 +25,8 @@ class P5Wrapper extends React.Component<IP5WrapperProps, IP5WrapperState> {
       ...props,
       canvas: null,
       // TODO: find a workaround since refs in state are bad practice
-      wrapper: this.wrapper
-    }
+      wrapper: this.wrapper,
```
jamesrweb: Can you update the config to prefer no trailing commas please?

jamesrweb (dismissed):
@yevdyko Looks good to me, some minor documentation additions and workflow updates to change but otherwise a really nice PR so far!

src/index.tsx
yevdyko: Done

.eslintignore
yevdyko: That's a good idea. I created a new PR #80 with these changes. Thanks for pointing this out!

.eslintrc
yevdyko: Yeah, as we don't have any other documentation yet, I have added it inside the `docs` folder.

.prettierrc
yevdyko: Done

package.json
yevdyko: Agreed, it's a good idea to configure CI for linting and testing. The only thing I'm concerned about is formatting files in CI with extra commits, I'm not sure this is the best approach. I think a pre-commit git hook would work better for that, but if you don't like it, I can move this step too.

.prettierignore
```diff
@@ -0,0 +1,5 @@
+dist
+node_modules
+package.json
+package-lock.json
+*.html
```
jamesrweb, line 5: Why do we ignore JSON and HTML, not an issue, just curious? 😄

package.json
jamesrweb: I really believe CI is the place for this, I updated the code myself, could you test my changes?

jamesrweb (commented):
@yevdyko Can you test my changes please?

I updated the CI and removed the git hooks because the CI is where we should be automating the integration of new code. Please check that I removed all non-necessary dependencies and that the CI workflow makes sense. On a sidenote, you mentioned extra commits, this would only happen if changes occur, not otherwise and so if linting and formatting don't change something, nothing is commited. The CI just helps the developer to develop and the CI to integrate so that the developer doesn't need to worry about code verification, formatting or stylistic issues and can write code how they wish and we format it to our needs, so long as it is reviewed / tested then it makes life easier for everyone in my experience plus it is what CI is for anyway.

Happy to answer any open questions and really happy with your work, this is really good stuff my friend! 👍🏻

.prettierignore
yevdyko: I guess JSON files can be removed, but they are still not included in the format script? Would you prefer to add them there?

As for HTML, it was excluded due to the default `printWidth` being 80, which I think is a bit narrow for them. Moreover, I don't think any changes will be needed there in the near future, and I have doubts that we even require these examples within the library.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+      if: ${{ failure() }}
+      run: npm run lint:fix
+
+    - name: Commit fixed lint issues
+      uses: stefanzweifel/git-auto-commit-action@v4.1.2
+      with:
+        commit_message: Apply fixed lint issues
+        branch: ${{ github.head_ref }}
+
+    - name: Run test suite
```
yevdyko: I would prefer the testing to be done in a separate job that is not blocked by lint issues.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,59 @@
...
+          ${{ runner.os }}-npm-
+        
+    - name: Lint files
+      run: npm run lint
+
+    - name: Linting failed (attempting fix...)
+      if: ${{ failure() }}
+      run: npm run lint:fix
+
+    - name: Commit fixed lint issues
```
yevdyko: We don't need to have both scripts (`lint:fix` and `format`) in CI since `format` is included in `lint:fix`, but the problem is that not all lint issues can be fixed automatically, so I would leave the only `format` step in the CI that can be always applied.

yevdyko:
@jamesrweb I don't mind having a formatting step in the CI, it's more a matter of taste :) Also, if it's inconvenient, we can always change the approach we take.

After you specified the master branch as a trigger for the push in the CI I can't test how it works together :)

~~Please add also changes in the package-lock.json file.~~ Committed the updated lock file.

Thanks for the feedback and support!

package.json
```diff
@@ -62,7 +69,7 @@
     "webpack-dev-server": "^3.11.0"
   },
   "peerDependencies": {
-    "p5": "*",
+    "p5": "^1.1.9",
```
jamesrweb, line 72: Do we need a specific version definition?

.github/workflows/ci.yml
jamesrweb: The CI flow should be blocked if code has lint issues and an automatic fix attempt fails. If it is not lintable, it is not testable after all. If you wish to propose a change, please update the PR 😄

.prettierignore
jamesrweb: I mean if we run a formatter anyway, what's the harm in just letting it format all our files (except node_modules, etc of course).

jamesrweb: Up to you, I'm not too bothered about this 👍🏻

.github/workflows/ci.yml
jamesrweb: > We don't need to have both scripts (lint:fix and format) in CI since format is included in lint:fix

> so I would leave the only format step in the CI that can be always applied

What do you mean by this? One runs prettier and the other eslint so I am confused at this statement, can you clarify?

> not all lint issues can be fixed automatically

Exactly which is why the `if: ${{ failure() }}` line is there, it will only ever execute this step if a failure happened in the previous `lint` step and if it runs `lint:fix` and it can't do it for some reason then it will throw an error at that step instead. The goal of this automatic fix attempt is to reduce developer load by allowing the integration suite to automate integration into the codebase as much as possible. If something is truly unfixable automatically then the dev must fix it but otherwise we can handle it for them here.

jamesrweb:
> @jamesrweb I don't mind having a formatting step in the CI, it's more a matter of taste :) Also, if it's inconvenient, we can always change the approach we take.
> 
> After you specified the master branch as a trigger for the push in the CI I can't test how it works together :)
> 
> ~Please add also changes in the package-lock.json file.~ Committed the updated lock file.
> 
> Thanks for the feedback and support!

I don't know why it does that and I cannot update the settings but the reason master is defined is because we only want the CI script to run for PR's into master or (if admin rights exist) pushed into master occur. If we had it run on PR A going into PR B which is targeted at master then in this case only PR B would run CI since A wouldn't directly need to since it wouldn't affect the main codebase. Does that make more sense? I amn't an expert so to say with GitHub actions so maybe I am missing something but I have found this flow useful in other personal projects but open to changes if we find a better way.

package.json
yevdyko: I would like to avoid updating the main dependency in the current PR and prefer to do it in a separate one.

.github/workflows/ci.yml
yevdyko: I have updated the CI configuration and moved the testing to a separate job that will run in parallel with the linting. Please see the changes in the last commit. Regarding the non lintable code is also not testable, I'm not sure about that - linting is about style and may not apply, but still preferable, and testing is about functionality that works as expected. For such a small library it's not critical to have these processes step by step, but for a long-running test suite, it's definitely better to have it in parallel.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,89 @@
...
+          CI: true
+
+      - name: Format files
+        run: npm run format
+
+      - name: Commit formatting changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Apply formatting changes
+          branch: ${{ github.head_ref }}
```
jamesrweb, line 45: Maybe this should be in a seperate format job instead of in the lint job?

jamesrweb (commented):
Only one thing left and then LGTM!

.github/workflows/ci.yml
yevdyko: > What do you mean by this? One runs prettier and the other eslint so I am confused at this statement, can you clarify?

When we use ESLint and Prettier together, we need to exclude the formatting rules in ESLint that might conflict with Prettier. For that, we use `eslint-config-prettier` and `eslint-plugin-prettier` libraries that run Prettier as ESLint rules and exclude conflict rules. Since we're not running Prettier just for JS and TS files, I changed the order and run the formatting step first and then the linting JS and TS files.

yevdyko:
> I don't know why it does that and I cannot update the settings but the reason master is defined is because we only want the CI script to run for PR's into master or (if admin rights exist) pushed into master occur. If we had it run on PR A going into PR B which is targeted at master then in this case only PR B would run CI since A wouldn't directly need to since it wouldn't affect the main codebase. Does that make more sense? I amn't an expert so to say with GitHub actions so maybe I am missing something but I have found this flow useful in other personal projects but open to changes if we find a better way.

Yes, I don't mind, and let's stick to this flow. Thanks for adding a step with cache npm packages! 🤖

.github/workflows/ci.yml
yevdyko: Since ESLint used Prettier as its rules, I think it can be done step by step. First formatting all selected file types, then linting JS and TS files. Let's see how it works in the next PRs, I'm only in favour of something to improve if it's going to be inconvenient.

jamesrweb: Can we try then? or do you want to merge as is?

yevdyko: The problem that can arise if we separate formatting and linting is that it will be parallel and not under the control of what happens in the first place. ESLint also runs the Prettier as rules so it can get messy That's why I think you can merge as it is and we can test it in the next PRs as it's not testable right now. Thanks!

PR #80, opened 2021-03-29
Build demo into dist folder

yevdyko:
This MR updates the Webpack config to build a demo into the `dist` folder.

#### What was changed?

- Format Webpack config.
- Build demo into the `dist` folder.
- Update `deploy` script with a new build directory path.
- Remove example build directory from ignore files.
- Use space indent in the `.editorconfig`.
- Prefer to use double quotes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "build": "npm run clean && tsc",
     "build-demo": "npm run clean && webpack --config webpack.config.demo.js --mode production",
     "clean": "rimraf dist",
-    "deploy": "gh-pages -d example/dist",
+    "deploy": "gh-pages -d dist",
     "publish-demo": "npm run build-demo && npm run deploy",
     "publish-npm-package": "npm run build && npm publish"
   },
--- a/webpack.config.demo.js
+++ b/webpack.config.demo.js
@@ -1,39 +1,39 @@
-const path = require('path');
-const HtmlWebpackPlugin = require('html-webpack-plugin');
+const path = require("path");
+const HtmlWebpackPlugin = require("html-webpack-plugin");
 const htmlWebpackPlugin = new HtmlWebpackPlugin({
-  template: path.join(__dirname, './example/src/index.html'),
-  filename: './index.html'
+  template: path.join(__dirname, "./example/src/index.html"),
+  filename: "./index.html"
 });
 
 module.exports = {
-  entry: path.join(__dirname, './example/src/app.jsx'),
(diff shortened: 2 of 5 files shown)


yevdyko:
@jamesrweb Could you test the deploy script?

package.json
```diff
@@ -72,7 +72,7 @@
     "build": "npm run clean && tsc",
     "build-demo": "npm run clean && webpack --config webpack.config.demo.js --mode production",
     "clean": "rimraf dist",
-    "deploy": "gh-pages -d example/dist",
```
jamesrweb, line 75: Run the `npm run clean` script here also because just like the publish script, we need a clean dist folder to be generated ❗

webpack.config.demo.js
```diff
@@ -1,38 +1,41 @@
...
+      }
     ]
-},
+  },
   plugins: [htmlWebpackPlugin],
   resolve: {
-    // Add '.ts' and '.tsx' as resolvable extensions.
-    extensions: [".ts", ".tsx", ".js", ".json"]
-},
+    extensions: ['.ts', '.tsx', '.js', '.json']
```
jamesrweb: Prefer double quotes in files for strings please 😄

jamesrweb (changes requested):
Looks good to me, just 2 minor points to change and I can merge once changed! Thanks for the awesome work man, you're on fire!👍🏻

webpack.config.demo.js
yevdyko: I can also update this setting in the formatter config in #79. Would you prefer double quotes in the whole project as it's not consistent at the moment?

package.json
yevdyko: Hmm, I'm not so sure about that, as we don't use this script separately, but only as part of the `publish-demo` script where we delete the `dist` folder at the `build` stage. Perhaps, we can remove the `deploy` script and be more specific in the `publish-demo`:

```
npm run build-demo && gh-pages -d dist
``` 

@jamesrweb What do you think about that?

webpack.config.demo.js
yevdyko: Done

jamesrweb: Yeah, ideally through the whole project but happy to merge this and accept the double quote change in #79 as it seems more appropriate there anyway.

package.json
jamesrweb: You're right, the scripts need updated and cleaned but for now this is working as expected. Good work!

jamesrweb (approved):

PR #82, opened 2021-04-10
Set up testing config

yevdyko:
This PR sets up a testing config incl. the initial render test and fixes the failed CI.

#### What was changed?

- Add jest and react testing libraries.
- Add jest config and script for running tests failed in CI.
- Add sketch fixture.
- Add initial render test.
- Add canvas as a dev dependency.
- Add names to jobs in Github actions.
- Fix fetching the repo from the forked branch.
- Run format and auto-correct steps only on push events.

@jamesrweb I think we can do proper testing in the next PRs and leave only a configuration here.


jamesrweb (changes requested):
Looks good but I would add at least a render test to be sure it renders properly and is Snapshot-able. Also the linter seems to be failing and cannot be auto fixed, can you look into that?

yevdyko:
> Also the linter seems to be failing and cannot be auto fixed, can you look into that?

Yeah, I see, but it looks like something went wrong on the Github side as it failed on the first step based on the error:

```sh
The process '/usr/bin/git' failed with exit code 1
```

I hope this goes away after restarting CI, but I can't find a manual re-run action so let's see if it works after adding a new commit with a render test.

.github/workflows/ci.yml
```diff
@@ -8,13 +8,14 @@ on:
...
   lint:
+    name: Lint
     runs-on: ubuntu-latest
 
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
         with:
-          ref: ${{ github.head_ref }}
+          ref: ${{ github.ref }}
```
jamesrweb, line 18: Why use `ref` over `head_ref`?

src/index.test.tsx
```diff
@@ -0,0 +1,5 @@
+describe("P5Wrapper", () => {
+  it("tests the component", () => {
+    // test case should be here
```
jamesrweb: Can you add a test here?

.github/workflows/ci.yml
yevdyko: `head_ref` value is only available for the PR created from the branch in the original repo, not the forked one. That's why CI has been failed with the error when tried to fetch the repo. We should test auto-commit steps afterward.

src/index.test.tsx
yevdyko: Yes, I can. Maybe, later in the evening. I need to figure out how to properly mock the canvas.

jamesrweb: We could have a sketch fixture for use across tests and then just render and assert on the render with testing library for example?

yevdyko: Sounds like a good idea. Thanks!

.github/workflows/ci.yml
yevdyko: FYI: This is the relevant [issue](https://github.com/actions/checkout/issues/366).

src/index.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import P5Wrapper from "./index";
+import sketch from "./__fixtures__/sketch";
+
+describe("P5Wrapper", () => {
+  xit("renders the component", () => {
```
yevdyko: @jamesrweb I don't know how to properly render the canvas component in the test and what to assert here. Could you help with that?

jamesrweb: @yevdyko added an initial render test, can you review and check why CI is failing?

package.json
```diff
@@ -36,45 +36,51 @@
     "url": "https://github.com/and-who/react-p5-wrapper/issues"
   },
   "dependencies": {
-    "p5": "^1.1.9"
+    "canvas": "^2.7.0",
```
yevdyko: @jamesrweb maybe, it should be a dev dependency?

As I understand it, we need the `canvas` package for testing. Or is it also used in the library itself?

src/index.test.tsx
yevdyko: @jamesrweb Thanks for your input!

I'm still looking into the problem in CI. Something is wrong with the reference to the forked branch. I need to double-check one solution in the test repository before asking for changes on your side.

package.json
jamesrweb: Yes, it should be a dev dependency, good catch 👌🏻

yevdyko: Okie, I'll move it to the right place :)

src/index.test.tsx
yevdyko: @jamesrweb After playing around with the test repo a bit and diving into the `git-auto-commit-action` documentation, I found that it doesn't work well with the forked repository due to Github actions restrictions. See the following details: [Using the Action in forks from public repositories](https://github.com/stefanzweifel/git-auto-commit-action#using-the-action-in-forks-from-public-repositories). It only applies changes on push commit after merging PR with red CI.

So I would like to avoid this behaviour and prefer to remove auto-correct steps from CI and show it failing if the changes don't pass the formatter or linter. What do you think about that?

jamesrweb: Sure but please reset the `ref` param on all instances to the `head_ref` then. Also we could create a separate workflow which only works when new code is pushed to master from local or a PR which does the formatting and fixes. Then the PR workflow just does linting and testing for example. What do you think?

yevdyko: Totally agreed, I've updated the current workflow based on your suggestions. This way we can use one workflow, but only run the formatting and auto-correct steps on the push events. We still need to use `ref` in checking out the repo step, otherwise, it won't work. So I think we can stop there for now and continue to improve the flow in the next iterations.

🚢 🚢

jamesrweb (approved):

Issue #70, opened 2020-12-06
Add support for React 17


disambiguator:
Can the package.json be updated to allow for React 17?

jamesrweb:
The package doesn't stop you using react 17, is there a specific issue you have seen or just an assumption that it wouldn't work?

jamesrweb:
I updated the dependencies in PR #71 and merged it. It will be released at the next possible point. For now though I see no reason you shouldn't be able to use the library as is? Happy to help if you can explain further though.

jamesrweb:
Closing for now.

disambiguator:
@jamesrweb Great thank you! I am seeing this when running `npm update`

```
❯ npm update
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: floating_points@1.0.0
npm ERR! Found: react@17.0.1
npm ERR! node_modules/react
npm ERR!   react@"^17.0.1" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.3.0" from react-p5-wrapper@2.3.0
npm ERR! node_modules/react-p5-wrapper
npm ERR!   react-p5-wrapper@"^2.3.0" from the root project
```

I figured there wouldn't actually be incompatible, thank you.

jamesrweb:
Peer dependencies are a tricky one with NPM, honestly I don't think we need them for the project but I don't want to remove them without feedback on why they were there in the first place.

For now I am the one working mostly on the repo and so all I can say is that in the future when v3 is complete (see initial PR #65) there will also be a tidying up of how dependencies, builds and configurations are handled. I plan to implement those in the next step once I fix the current re-rendering issue in that PR.

I need @and-who to do the NPM releases of the codebase (which includes quite a few changes already) since I don't have permissions to do so but he is quite busy currently so I hope he will be able to do this sometime this month but you will be notified once it is done either way.

Hopefully that will resolve the issue once and for all but for now I have done all I can in terms of altering the package.json for you.

disambiguator:
I really appreciate it. I've never published my own package and have been curious how to keep these things up to date in a hassle-free way, so learning a lot here. :) Thanks for maintaining this!

PR #85, opened 2021-05-14
Update dependencies to the latest versions

yevdyko:
This MR updates dependencies to the latest versions and removes `p5` from the peer dependencies.

#### What was changed?

- Update `p5` version to the latest 1.3.1.
- Update dependencies to the latest versions.
- Remove `p5` from peer dependencies.


jamesrweb:
@yevdyko Do we need the dependency in both dependencies and peer dependencies?

Also, does the .x syntax work? Never seen this before and if it does, can you give a quick explainer for future reference?

yevdyko:
@jamesrweb I think you are right about `p5` as a dependency and we should remove it from the peer dependencies as it makes no sense.

As for the `1.x` syntax, it's similar to `^1.0.4`. See more details here: [Using semantic versioning to specify update types your package can accept](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept)

Maybe, we need to add the minimum version for React as a peer dependency based on the used features in the library.

jamesrweb:
> @jamesrweb I think you are right about `p5` as a dependency and we should remove it from the peer dependencies as it makes no sense.
> 
> As for the `1.x` syntax, it's similar to `^1.0.4`. See more details here: [Using semantic versioning to specify update types your package can accept](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept)
> 
> Maybe, we need to add the minimum version for React as a peer dependency based on the used features in the library.

After some thinking, I remembered that it was actually added there by me as a peer dependency because of [an issue I resolved](https://github.com/and-who/react-p5-wrapper/issues/70#issuecomment-739559298) where users installing the library has a `p5 is not defined` or a related error appearing (this happened because @and-who [removed the peer dependency previously](https://github.com/and-who/react-p5-wrapper/issues/29#issuecomment-498144050)) and so I re-added the peer dependency in #71. Can you test locally using `npm link` to see if the install throws such an error again now? Maybe safer to revert it to exist to stop a repeat of the issues in #70, not sure, what do you think?

I would prefer using the `^x.x.x` syntax since the other packages are like that, what do you think, should we revert or update all, if so, is there a benefit or just a preference? Happy either way but would like to keep things consistent.

yevdyko:
> After some thinking, I remembered that it was actually added there by me as a peer dependency because of an issue I resolved where users installing the library has a p5 is not defined or a related error appearing (this happened because @and-who removed the peer dependency previously) and so I re-added the peer dependency in #71. Can you test locally using npm link to see if the install throws such an error again now? Maybe safer to revert it to exist to stop a repeat of the issues in #70, not sure, what do you think?

I removed the `p5` from peer dependencies and double-checked in the demo application locally using `link` that everything works properly until we have `p5` as a dependency. I think this was the main reason why users were having some problems. You added `p5` to dependencies on [23 Sep 2020](https://github.com/and-who/react-p5-wrapper/commit/9ddae1801e373f4dcac5fe857ed4496387509260), but it has not been published in `npm` and `yarn` until [14 Mar 2021](https://github.com/and-who/react-p5-wrapper/releases/tag/v2.4.0). As a result, users were forced to install `p5` separately as it was in the peer dependencies but not installed while the package installation. So we can now safely remove it from the peer dependencies and leave it as a dependency only.

> I would prefer using the ^x.x.x syntax since the other packages are like that, what do you think, should we revert or update all, if so, is there a benefit or just a preference? Happy either way but would like to keep things consistent.

The only reason to use `1.x` was to have `p5` as a peer dependency. I thought it would be more obvious which versions are acceptable, but once we move away from that, we can use the common `^x.x.x.` syntax.

jamesrweb:
Merged the current PR, for follow up topics we can open a new one @yevdyko, great work so far, appreciate the contributions! 👌🏻

PR #93, opened 2021-06-19
Remove redundant dotfiles

### LINKED ISSUE #88: Investigate opportunities to remove dotfiles and integrate things into the CI
@yevdyko

yevdyko:
This MR removes the redundant dotfiles that can be replaced by using configs in the `package.json` file.

We can remove the `.npmignore` file because we control the included files in the published package through the `files` property in the `package.json` file. The `.prettierrc` and `.eslintignore` configs are so minimal that can be easily moved to the `package.json` file too. The rest of the dotfiles are fine for me.

### What was changed?

- Move eslintignore config to the package.json file.
- Move prettier config to the package.json file.
- Remove .npmignore file.


jamesrweb:
Changed the base branch so that this can be a part of the v3 refactor, I think the next tasks we complete should have their base branch set to there instead of `master` until we complete the v3 setup in the coming week(s) and then we can release them in one go for version 3.

jamesrweb (approved):

PR #94, opened 2021-06-19
Revert "Create .gitattributes"

yevdyko:
This MR reverts the commit that added git attributes settings because it does not allow us to control changes to the `package-lock.json` file using git.

```shell
$ git diff package-lock.json

diff --git a/package-lock.json b/package-lock.json
index 671909d..2e466c6 100644
Binary files a/package-lock.json and b/package-lock.json differ
```

I don't want to work out how to set this up properly and would prefer to undo the change for now and add these settings meaningfully later if necessary.

### What was changed?

- Revert "Create .gitattributes".


jamesrweb:
I don't feel a reversion is the answer, the issue here is that the attributes file hasn't got a `difftool` defined from what I gather from your screenshot, in this case you can try [this](https://superuser.com/a/1449340) or extend the gitattributes to have a default setting for such files, in the same thread as the previous link you will find simple solutions to this also. If you still have issues, we can discuss that in our next meeting, WDYT? 👍🏻? 👎🏻?

jamesrweb:
I also wonder why you want to diff the lock file rather than the package.json? The lock is only for npm internal things.

jamesrweb:
Closed for now based on our previous meeting. You can always update the attributes file to allow diffing of the lockfile as discussed in a followup PR.

PR #100, opened 2021-06-23
Update integration scripts

jamesrweb:
This PR fixes the failing CI on master by adding a workflow personal access token into the repository secrets for use when checking out the code inside actions. I also split the responsibilities for what is required when master is pushed to (on merge) and a PR is submitted (on review).


.github/workflows/PR.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - uses: c-hive/gha-npm-cache@v1
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Lint
+        run: npm run lint
+      
+      - name: Linting failed (attempting fix...)
```
yevdyko, line 27: I think this step can be removed for now as it makes no sense without the git-auto-commit action.

.github/workflows/Push.yml
```diff
@@ -0,0 +1,31 @@
...
+        run: npm ci
+
+      - name: Format
+        run: npm run format
+
+      - name: Commit formatting changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Apply formatting changes
+          branch: ${{ github.head_ref }}
```
yevdyko, line 31: A new line at the end of the file is missing.

.github/workflows/Push.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+
+      - uses: c-hive/gha-npm-cache@v1
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Format
```
yevdyko, line 24: Does it mean no more commits directly to the master branch unless we run tests and linter?

yevdyko (dismissed):
Looks good! I just noticed a redundant step in the PR workflow.

Thanks @jamesrweb!

🚢

.github/workflows/PR.yml
jamesrweb: True, will update and remove later today, good catch!

.github/workflows/Push.yml
jamesrweb: No, this file only runs when a PR is merged into to `master` so commits will still happen but formatting won't occur in the PR on push many times, only one time when the PR is merged!

.github/workflows/Push.yml
jamesrweb: Why do we need an empty new line? Either way when code is merged the format script will run and format the code when it's in master in the way it is configured to in the `.prettierrc`

jamesrweb: Talking about the prettier config, I will make some updates to that because it currently just has one rule 🤨. Will push that here later too!

.github/workflows/Push.yml
yevdyko: I mean, when you send a commit directly to the master without creating a PR, the CI will only do the formatting and skip the linting and testing steps because they are not included in the workflow. So my expectation is that we will only make changes through the PR based on these settings for CI.

jamesrweb: Exactly, remember the branch rules we have? Master can only be pushed to via PR now, even force pushes aren't allowed to master.

yevdyko: That makes sense! Sorry, I didn't remember these settings, so I decided to clarify.

.github/workflows/Push.yml
yevdyko: Regarding the prettier configuration, I think it was your preference to avoid trailing commas, which is different from the default.

jamesrweb: Yeah but we can do more, I will update it in a different PR actually

jamesrweb: Updated the prettier config in PR #102

.github/workflows/PR.yml
jamesrweb: Just added a commit action for this actually since if the fixer fails CI won't commit anyway but if it passes, we don't pass it to the user to fix locally and then re-push. Should make it easy for contributors to get it right more often than not since most issues would be minor styling topics, etc I would think.

yevdyko (approved):
🚀

PR #101, opened 2021-06-23
Convert to functional style and expose more types for the upcoming version 3 release

### LINKED ISSUE #87: Add working example to `refactor/v3` branch

### LINKED ISSUE #90: Investigate flickering issue on `refactor/v3` branch

jamesrweb:
Closes #90.
Closes #87.

This PR will, when completed, implement a new functional react version of the wrapper instead of the old class based component. The docs need updated, some build configurations need changed to work with `rollup` as we move away from `webpack` but overall the implementation is nearly complete!


package.json
```diff
@@ -20,85 +45,62 @@
...
   },
   "bugs": {
     "url": "https://github.com/jamesrweb/react-p5-wrapper/issues"
   },
+  "dependencies": {
+    "deep-equal": "^1.1.1",
+    "p5": "^1.3.1"
+  },
+  "peerDependencies": {
+    "react": "*",
```
yevdyko, line 62: I'm wondering if you want to define the React version here, based on the fact that you use React hooks in your implementation. For example, `16.8` is the version that supports them.

src/index.tsx
```diff
@@ -1,85 +1,50 @@
...
+export interface Sketch {
+  (instance: p5): void;
 }
 
-export interface IP5WrapperState {
-  sketch: (p: p5) => void;
-  attributes: TAttributes;
-  canvas: p5;
-  wrapper?: HTMLElement;
+export interface P5WrapperProps {
```
yevdyko, line 9: After changing the component name to `ReactP5Wrapper` should it also be applied here? `ReactP5WrapperProps`?

src/index.tsx
```diff
@@ -1,85 +1,50 @@
...
-export default P5Wrapper;
+const ReactP5WrapperComponent: FC<P5WrapperProps> = ({
+  sketch,
+  children,
+  ...props
+}) => {
+  const wrapper = createRef<HTMLDivElement>();
+  const [instance, setInstance] = useState<P5Instance>();
+
+  useEffect(() => {
```
yevdyko, line 30: 🍬

yevdyko (approved):
Outstanding work!

It would be nice if you could explain in the description the difference between v2 and v3, why you introduced these changes and what problems do they solve?

yevdyko:
Auto-merge was not the behaviour I expected 😕

PR #102, opened 2021-06-25
Update prettier config and update code

jamesrweb:


.prettierrc
```diff
@@ -1,3 +1,19 @@
 {
-  "trailingComma": "none"
+	"arrowParens": "avoid",
+	"bracketSpacing": true,
```
yevdyko: I think the following parameters do not make sense to include in the configuration because they have default values:

- `tabWidth`
- `useTabs`
- `semi`
- `requirePragma`
- `quoteProps`
- `printWidth`
- `jsxBracketSameLine`
- `insertPragma`
- `htmlWhitespaceSensitivity`
- `embeddedLanguageFormatting`
- `bracketSpacing`
- `vueIndentScriptAndStyle`

package-lock.json
```diff
@@ -1,18007 +1,8 @@
 {
   "name": "react-p5-wrapper",
   "version": "2.4.1",
-  "lockfileVersion": 2,
```
yevdyko: I assume that the `package-lock.json` file should not be included in the PR because there were no changes in the packages.

.github/workflows/ci.yml
```diff
@@ -51,11 +49,10 @@ jobs:
         run: npm run lint
 
       - name: Linting failed (attempting fix...)
-        if: ${{ failure() && github.event_name == 'push' }}
```
yevdyko: These changes will conflict with the existing PR #100

README.md
```diff
@@ -54,19 +58,21 @@ export default App;
...
 
-In the below example you see the `myCustomRedrawAccordingToNewPropsHandler` function, which is called when the properties of a wrapper component are changed.
+In the below example you see the `myCustomRedrawAccordingToNewPropsHandler`
+function, which is called when the properties of a wrapper component are
+changed.
 
 ```js
-import React from "react";
-import P5Wrapper from "react-p5-wrapper";
+import React from 'react';
```
yevdyko: What is the reason to change the formatting again after we agreed on double quotes months ago? 

I'm wondering why you're redoing everything I've already done. There is no improvement or value here. Perhaps, I'm missing something.

yevdyko (changes requested):

.prettierrc
jamesrweb: I know they are but if prettier changes their defaults or someone wants to know what we use or they could change, I think it is better to be explicit, WDYT?

README.md
jamesrweb: So the quotes we could use doubles but prettier defaults on most other file types to singles and so for consistency with other file types I figured it would be better to stick with that, WDYT?

package-lock.json
jamesrweb: Lock files should always be committed but not sure why the version key is gone, doesn't matter much since the package.json is all that matters, this file is just used by npm itself anyway and no dependencies changes so if npm changed this, I am sure there is a reason, right?

.github/workflows/ci.yml
jamesrweb: This isn't an issue since this file doesn't exist now.

README.md
yevdyko: We've already done it the other way around here: https://github.com/jamesrweb/react-p5-wrapper/pull/79/commits/ca0655101544954839df03e85945284bdcff937b

jamesrweb: If you really want double quotes then I can set it back, it makes no difference at the end of the day so it's up to you?

.prettierrc
yevdyko: This contradicts the fact that you want to minimize the configuration files, I would prefer to close this PR as I am happy with the current configuration and state of the library, the package is so small but you're now starting to rewrite everything to your preference, which makes no sense.

README.md
jamesrweb: Updated 🤷🏻‍♂️

.prettierrc
jamesrweb: We are aiming to reduce the dotfiles, not the configuration. #93 deals with that or am I missing something that we previously discussed?

yevdyko:
@jamesrweb I'm wondering if you can align these changes with the new file structure to avoid conflicting files.

jamesrweb:
@yevdyko rebased master and moved configuration to the config directory.

yevdyko (approved):
🚀

PR #106, opened 2021-06-28
Cleanup configurations and update dependencies

### LINKED ISSUE #104: Auto-formatting doesn't work on push to master
@jamesrweb There is an error in CI when running git-auto-commit action:

```
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: At least 1 approving review is required by reviewers with write access.        
To https://github.com/jamesrweb/react-p5-wrapper
 ! [remote rejected] master -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/jamesrweb/react-p5-wrapper'
```

https://github.com/jamesrweb/react-p5-wrapper/runs/2925875088

jamesrweb:
## What changed?

- Added github specific files for support, pull request and issue templates for example
- Exposed new `SketchProps` type so that consumers can have a type rather than `any` to set some form of expectation of what they should do when using the `updateWithProps` function within a sketch
- Move configuration files into their own directory
- Cleaned up some files and inlined others
- Updated project scripts to use the new config locations
- Updated packages to latest versions to fix security issues

It is worth noting that this closes #104 when merged since the format commit step won't need to be done on merge anymore and the linter will attempt fixes instead so that the master branch doesn't have the double CI trigger issue anymore.


yevdyko (approved):
🚢

PR #108, opened 2021-07-12
Update tests

jamesrweb:
## Proposed Changes

- Update tests to cover client and server rendering cases
- Update tests to cover prop updates


src/index.tsx
```diff
@@ -7,7 +7,7 @@ export interface SketchProps {
 }
 
 export interface Sketch {
-  (instance: p5): void;
+  (instance: P5Instance): void;
```
jamesrweb, line 10: This makes sure that when users using typscript, for example, want to implement the `updateWithProps` function then proper typings for that callback will be applied without the need for manual typing. So instead of:

```ts
p5.updateWithProps((props: SketchProps) => {});
```

Someone could just do:

```ts
p5.updateWithProps(props => {});
```

and the types should be correctly inferred.

src/index.tsx
```diff
@@ -30,18 +30,17 @@ const ReactP5WrapperComponent: FC<P5WrapperProps> = ({
...
   useEffect(() => {
     if (wrapper.current === null) return;
     instance?.remove();
     const canvas = createCanvas(sketch, wrapper.current);
-    canvas.updateWithProps?.(props);
     setInstance(canvas);
   }, [sketch, wrapper.current]);
 
+  useEffect(() => {
+    instance?.updateWithProps?.(props);
```
jamesrweb, line 41: Since `useEffect` calls run in order, we don't need the second call that used to exist which means we have one line of code less and deduplicated logic which works for me any day of the week 😉!

tests/helpers/streams.ts
```diff
@@ -0,0 +1,8 @@
+export function unwrapReadableStream(stream: NodeJS.ReadableStream) {
```
jamesrweb, line 1: This helper allows us to unwrap readable streams to get to their contents. We need this for the server rendering tests.

src/index.tsx
yevdyko: This is a good thing. I guess that was done this way in the previous version. 

👍

src/index.tsx
yevdyko: Makes sense, thanks for this!

yevdyko (approved):
LGTM @jamesrweb

Thanks for the clarification on the code!

🚢

PR #115, opened 2021-09-10
Updating broken example in readme

jamesrweb:
One of the examples in the readme is broken, I updated the docs to fix this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,11 +61,9 @@ function sketch(p5) {
   };
 }
 
-function App() {
+export default function App() {
   return <ReactP5Wrapper sketch={sketch} />;
 }
-
-export default App;
 ```
 
 ### Props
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -111,6 +111,16 @@ function sketch(p5) {
 }
 
 function App() {
+  const [rotation, setRotation] = useState(0);
+
+  useEffect(() => {
+    const interval = setInterval(rotation => setRotation(rotation + 1), 100);
```
jamesrweb: @yevdyko maybe this isn't a good example but it fixes the bug, we could replace this with something else, WDYT?

yevdyko: It's still broken:
- `useState` and `useEffect` must be imported from `React`. 
- `rotation` is `NaN` in props.

yevdyko (changes requested):
@jamesrweb can you double-check the fix?

README.md
jamesrweb: Updated the example and tested locally 👍🏻

jamesrweb:
@yevdyko can you re-read and re-test please, thanks! 🙏🏻

yevdyko (approved):
Looks awesome! Thanks @jamesrweb for the fix!

🚢

jamesrweb:
> Looks awesome! Thanks @jamesrweb for the fix!
> 
> 
> 
> 🚢 

🥳

PR #133, opened 2022-01-02
Bundle size improvements

jamesrweb:
With these changes we see a considerable reduction in bundle size.

## Proposed Changes

- Swap `deep-equal` for `microdiff`
- Remove comments in final bundle

## Notes

With these two changes we see a marked reduction in bundle size, here are my local benchmarks:

| Bundle | Original size (bytes) | New size (bytes) | Difference (%) |
| ------ | --------------------- | ---------------- | -------------- |
| ESM    | 1768                  | 958              | ~46% reduction |
| CJS     | 1948                  | 1137             | ~42% reduction |

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -13,7 +13,11 @@ const plugins = [
     typescript: require("typescript"),
     tsconfig: join(__dirname, "tsconfig.json")
   }),
-  terser()
+  terser({
+    format: {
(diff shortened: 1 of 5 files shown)


yevdyko (approved):
Awesome improvements!

Thanks for taking care about this @jamesrweb 🚀

PR #134, opened 2022-01-10
Use gitignore to avoid duplication in prettier ignore

jamesrweb:
## Proposed Changes

- Remove .prettierignore from project root
- Utilise the .gitignore in its place


yevdyko (approved):
Yeah, that's a good idea!

Thanks for implementing it!

🐑

PR #135, opened 2022-01-20
Update dependencies

jamesrweb:
## Proposed Changes

- Update project dependencies to latest versions
- Allowed legacy peer dependencies in the lock file temporarily
- Resolves the 3 outstanding security vulnerabilities found within dependencies by Dependabot


yevdyko (changes requested):
Good stuff!

Thanks for the upgrade @jamesrweb 🚀

yevdyko (approved):

PR #136, opened 2022-01-20
Configuration cleanup

jamesrweb:
## Proposed Changes

- Have a unified build to tidy the release script
- Alter the Webpack config to hold all possible variables rather than passing via the CLI
- Move CLI params into the ESLint config and set the ignore path via the flag instead of within the package file


yevdyko (dismissed):
Looks good! I've tested locally and all the scripts work as expected.

Thanks for this improvement @jamesrweb 🚀

yevdyko (approved):

Issue #127, opened 2021-10-19
Add example of abstracted setup and draw functions


jamesrweb:
We should document the use case that came up during the resolution of #126. 

This should show how to abstract the setup and draw functions and provide a small example setup.

PR #137, opened 2022-01-23
Add new example of declaring sketch

yevdyko:
This MR improves the README file by adding a new example of declaring a sketch in JavaScript.

Closes issue #127

## Proposed Changes

- Add an example of declaring a sketch using abstracted setup and draw functions.
- Use the same example in JavaScript and TypeScript sketches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Then just open `http://localhost:3001` in a browser.
 
 ### Javascript
 
+#### Option 1: Declaring a sketch as a function
+
 ```js
 import React from "react";
 import { ReactP5Wrapper } from "react-p5-wrapper";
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,6 +70,41 @@ export default function App() {
 }
 ```
 
+#### Option 2: Declaring a sketch using abstracted setup and draw functions
```
jamesrweb: Since this is more of a technique rather than a JavaScript specific thing, perhaps it should be in a separate area since some users might think this isn't possible in typescript for example. I'm not sure what to name the section though, something like "Notes" or "FAQs" (maybe not correct since there would only be this one answer but could be appropriate) or "Techniques"... I don't know, maybe none of those, What do you think?

README.md
```diff
@@ -68,6 +70,41 @@ export default function App() {
...
+    p5.pop();
+  };
+}
+
+function sketch(p5) {
+  p5.setup = setup(p5);
+  p5.draw = draw(p5);
+}
+
+export default function App() {
```
jamesrweb: Maybe we want to also update the readme to stop exporting as `default` in the examples and instead prefer named exports since we also do that in the source code for the wrapper component. It's not strictly related to the PR but what do you think of this?

yevdyko: Sounds reasonable - updated.

README.md
yevdyko: Agreed, let's move this to `Techniques` or something like that if I can find a more appropriate title.

jamesrweb: Sounds good, let me know what you're thinking and once it's pushed we can move this one forward for the next release :)

yevdyko: In general, I moved "Move using abstracted setup and draw functions" to a separate section as part of the "Usage". Also used the same example for JavaScript and TypeScript sketches. @jamesrweb What do you think?

jamesrweb: Looks okay but why is there a babel configuration there?

yevdyko: By mistake - already removed.

jamesrweb: Looks good to me 👌🏻

jamesrweb: Wait, the typescript examples are wrong now? The types aren't correctly imported, etc

jamesrweb: Did you also mean to make all the examples exactly the same?

yevdyko: I removed the props from them, but their types stayed the same.

yevdyko: Got it - let me quickly fix the types.

yevdyko: @jamesrweb Please verify this again?

yevdyko: > Did you also mean to make all the examples exactly the same?

Yes, the idea is to use the same example in JavaScript and TypeScript sketches. Previously, we used a sketch with props in TypeScript. We still have it in the Props section.

yevdyko: @jamesrweb I'm done with the changes

jamesrweb (approved):
Seems ok to me 🤞🏻

PR #145, opened 2022-02-13
Resolve unmounting instance issue

### LINKED ISSUE #142: P5 instances not remove()'d on component unmount?
Hi there!

Thanks for a nifty little wrapper! 🙏🏼 

I'm having some issues in my Next.js setup however, the P5 instance keeps running when I navigate between pages. I may very well just have missed something in my implementation for the below issue - but I couldn't figure out what 🤷🏼.

NB. I import using next/dynamic ReactP5Wrapper to avoid SSR:
```javascript
const ReactP5Wrapper = dynamic(
  () => {
    return import("react-p5-wrapper")
      .then((m) => m.ReactP5Wrapper)
      .catch(console.error);
  },
  { ssr: false }
);
```

## Expected Behavior
When a component including a `<P5Wrapper>` is unmounted, I would expect the P5 instance to be cleaned up and removed (with `remove()`). 

## Actual Behavior
The P5 instance continues to live and run even though the component is unmounted.

Here's a codesandbox demonstrating the issue: https://codesandbox.io/s/prod-architecture-tf8l0 

1. Navigate to "Page with ReactP5Wrapper", containing the P5 example.
3. Navigate back the landing page again and notice components on page B are unmounted but the P5 instance is still running (the console.log from within the sketch's `draw()` keeps running when using the browser's back button).

## Workaround / suggested solution
I've worked around it by wrapping the wrapper (....🤪) and keeping tabs on the p5 instance with a ref, and removing it on unmount:

```javascript
const SketchWrapper = ({ sketch }) => {
  const p5InstancRef = useRef();

  useEffect(() => {
    return () => {
      if (p5InstancRef.current) {
        p5InstancRef.current.remove();
     }
    };
  }, []);

  return (
    <ReactP5Wrapper
      sketch={(p5) => {
        p5InstancRef.current = p5;
        sketch(p5);
      }}
    />
  );
};
```

Package Version: `3.1.0`

jamesrweb:
Fixes issue #142 

## Proposed Changes

- Adds a new effect handler to remove the instance when the component is unmounted
- Update the dependencies to their latest versions
- Update the example to test mount and unmount behaviours
- Added a test to check that when the component is unmounted, the containing element has no `innerHTML` content. This in turn triggers the unmount effect.


src/index.tsx
```diff
@@ -39,11 +39,13 @@ const ReactP5WrapperComponent: FC<P5WrapperProps> = ({
...
     setInstance(canvas);
-  }, [sketch, wrapper.current]);
+  }, [sketch]);
 
   useIsomorphicEffect(() => {
     instance?.updateWithProps?.(props);
-  }, [props]);
+  }, [props, instance]);
+
+  useIsomorphicEffect(() => () => instance?.remove(), []);
```
fredrikelmegren, line 48: Haven't had time to test it out, but this here line looks like it'd indeed solve #142 😃

fredrikelmegren (approved):

yevdyko (approved):
Thanks for the extra testing @jamesrweb!

Looks good to me 🐑

PR #153, opened 2022-03-21
Update instance removal logic on component unmount

### LINKED ISSUE #142: P5 instances not remove()'d on component unmount?

jamesrweb:
Fixes #142

## Proposed Changes

- Use refs to track the instance instead of state
- Use types over interfaces for consistency

## Additional notes

Please await a review from @jhrtn before merging.

Relates to the #145

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,7 +1,6 @@
 import diff from "microdiff";
 import p5 from "p5";
-import React, { createRef, FC, memo, MutableRefObject } from "react";
-import { useRef } from "react";
+import React, { createRef, FC, memo, MutableRefObject, useRef } from "react";
 import { useIsomorphicEffect } from "rooks";
 
 type Wrapper = HTMLDivElement;
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,28 +1,36 @@
 import diff from "microdiff";
 import p5 from "p5";
-import React, { createRef, FC, memo, useState } from "react";
+import React, { createRef, FC, memo, MutableRefObject } from "react";
+import { useRef } from "react";
 import { useIsomorphicEffect } from "rooks";
 
 type Wrapper = HTMLDivElement;
-
-export interface SketchProps {
+export type Sketch = (instance: P5CanvasInstance) => void;
+export type SketchProps = {
   [key: string]: any;
-}
-
-export interface Sketch {
-  (instance: P5Instance): void;
-}
-
-export interface P5WrapperProps extends SketchProps {
+};
+export type P5WrapperProps = SketchProps & {
   sketch: Sketch;
-}
-
-export interface P5Instance extends p5 {
+};
+export type P5CanvasInstance = p5 & {
   updateWithProps?: (props: SketchProps) => void;
+};
+
+// @TODO: remove in next major version, keep for compatibility reasons for now.
+export type P5Instance = P5CanvasInstance;
+
+function createCanvasInstance(
+  sketch: Sketch,
+  wrapper: Wrapper
+): P5CanvasInstance {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -1,53 +1,61 @@
 import diff from "microdiff";
 import p5 from "p5";
-import React, { createRef, FC, memo, useState } from "react";
+import React, { createRef, FC, memo, MutableRefObject } from "react";
+import { useRef } from "react";
```
jhrtn:
```suggestion
import React, { createRef, FC, memo, MutableRefObject, useRef } from "react";
```

jhrtn (commented):
Thanks for the prompt update! I've tested this in my minimal sandbox, the official example and the app I'm using the lib for and it appears to resolve the issue in all cases

src/index.tsx
yevdyko: good suggestion @jhrtn - thanks for that!

src/index.tsx
```diff
@@ -1,53 +1,60 @@
...
+}
+
 const ReactP5WrapperComponent: FC<P5WrapperProps> = ({
   sketch,
   children,
   ...props
 }) => {
   const wrapperRef = createRef<Wrapper>();
-  const [instance, setInstance] = useState<P5Instance>();
+  const instanceRef = useRef<P5CanvasInstance>();
```
yevdyko: What about naming unification after introducing the `canvasInstanceRef` parameter in the `removeCanvasInstance` and using `canvasInstanceRef` instead of `instanceRef`?

yevdyko (changes requested):
@jamesrweb I've tested the fix using a minimal sandbox from @jhrtn and everything works as expected.

The only thing that can be improved is the unification of the mentioned variable name.

src/index.tsx
jamesrweb: Updated the naming, good idea 😄!

yevdyko (approved):
Great job @jamesrweb! 🚢

PR #157, opened 2022-03-27
Update CI workflow and configurations

jamesrweb:
## Proposed Changes

- Update the CI workflow to handle formatting just the same as we do for linting
- Rename the CI workflow file to reflect the configuration name more correctly
- Update the rollup configuration to more dynamically generate the bundle configurations
- Create new npm scripts for a local ci reproduction case and for checking formatting rules are being applied correctly


yevdyko (approved):
Thanks @jamesrweb, this looks good to me! 🚢 🚢 🚢

PR #166, opened 2022-05-29
Updates to docs, examples and tests.

### LINKED ISSUE #164: Render children on top of canvas
## Expected Behavior
To be able to render react components on top of the canvas

## Actual Behavior
If I use
``` js
const App = () => {
  return (
    <ReactP5Wrapper sketch={sketch}>
      <div>Hello</div>
    </ReactP5Wrapper>
  );
};
```
The canvas is simply rendered after the `<div>`

1. Follow tutorial on readme
2. Add child components

Package Version: `^3.3.0`

jamesrweb:
Fixes #164.

## Proposed Changes

- Update README to show expected usage of the `P5CanvasInstance` over the `P5Instance` type since this will be deprecated in the next major version.
- Update the code for the example to be more efficient and clean.
- Add a class to the wrapper for users who wish to target the wrapper or child elements with css. This was of course possible but the change here will make it "easier".

Fixes, if you can call it that, #164. I say "if you can call it that" since as I mentioned before, using css or js was previously possible for such use cases and this just makes it a little simpler with a class to specifically target other than P5s own one on the canvas proper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,17 +74,16 @@ TypeScript sketches can be declared in two different ways, below you will find
 two ways to declare a sketch, both examples do the exact same thing.
 
 In short though, the `ReactP5Wrapper` component requires you to pass a `sketch`
-prop. The `sketch` prop is typed as a `(instance: P5Instance): void;`. As long
-as the function declaration of your sketch is set to take in a single argument
-of type `P5Instance`, you are good to go!
+prop. The `sketch` prop is simply a function which takes a `p5` instance as it's
+first and only argument.
 
-#### Option 1: Declaring a sketch using the `P5Instance` type
+#### Option 1: Declaring a sketch using the `P5CanvasInstance` type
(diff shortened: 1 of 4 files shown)


jamesrweb:
@yevdyko please ensure you test the documentation examples and the example app itself as part of the review. Thanks in advance 🚀!

yevdyko:
> @yevdyko please ensure you test the documentation examples and the example app itself as part of the review. Thanks in advance 🚀!

@jamesrweb the changes look correct in the diff, tomorrow I will check the code in the examples so that it works as expected. Thanks for the improvement and hopefully it will be easier to use the component in the future 💪

yevdyko (approved):
@jamesrweb I've tested the typescript examples from the README as well as the demo application. Everything works as expected, as far as I can see.

🚢 🚢

PR #169, opened 2022-06-19
Update major changes in dependencies

jamesrweb:
## Proposed Changes

- Update all dependencies to latest major versions.
- Alter jest config to support their new environment loading style.
- Remove NodeStream tests for react server rendering as these APIs are currently and / or will be deprecated, [as seen in the react docs](https://reactjs.org/docs/react-dom-server.html#rendertonodestream).
- Update supported peer dependency of react to be the latest version of React 17 to encourage incremental upgrading now that version 18 is out.


yevdyko (approved):
@jamesrweb Thanks for cleaning up the configuration files and tests, adding new jest settings and updating dependencies. Everything looks good!

🚢

PR #171, opened 2022-07-25
Fix import name in example

jamesrweb:
## Proposed Changes

- Fix import name in children styling example


yevdyko (approved):
Good catch! Thanks @jamesrweb!

🛥️

PR #174, opened 2022-07-25
Add automated deployments of the demo application on GH Pages

jamesrweb:
## Proposed Changes

- Automates the deployments of the GH Pages demo application


example/index.html
```diff
@@ -1,17 +1,35 @@
 <!DOCTYPE html>
-<head>
-  <title>react-p5-wrapper</title>
+<head lang="en">
+  <meta charset="UTF-8" />
```
jamesrweb, line 3: Add meta tags to support older browsers a bit better with responsiveness, etc.

example/index.html
```diff
@@ -1,17 +1,35 @@
...
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>@P5-wrapper/react</title>
```
jamesrweb, line 6: Change project name to match repo.

example/index.html
```diff
@@ -1,17 +1,35 @@
...
 </head>
 <body>
   <div class="container">
-    <h1>react-p5-wrapper</h1>
+    <h1>@P5-wrapper/react</h1>
     <h2>
-      <a href="https://github.com/P5-wrapper/react">View project on GitHub</a>
+      <a
+        href="https://github.com/P5-wrapper/react"
+        target="_blank"
```
jamesrweb, line 14: Update links to open in a new tab with security measures using `rel` added.

example/index.html
yevdyko, line 6:
```suggestion
  <title>P5-wrapper/react</title>
```

Do we really need `@` here?

yevdyko: I'm not sure about this change, since the package is still called `react-p5-wrapper`? WDYT?

example/index.html
```diff
@@ -1,17 +1,35 @@
...
+<head lang="en">
+  <meta charset="UTF-8" />
+  <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>@P5-wrapper/react</title>
 </head>
 <body>
   <div class="container">
-    <h1>react-p5-wrapper</h1>
+    <h1>@P5-wrapper/react</h1>
```
yevdyko, line 10:
```suggestion
    <h1>P5-wrapper/react</h1>
```

example/index.html
yevdyko: 👍

.github/workflows/CI.yml
```diff
@@ -1,18 +1,22 @@
...
 
 on:
   pull_request:
     branches: [master]
 
+concurrency: ci-${{ github.ref }}
+
 jobs:
-  validate:
+  format-lint-test-and-build:
```
yevdyko, line 10:
```suggestion
  build:
```
What about just calling it `build`?

I think that's a bit better than repeating the step names in the job name, and the meaning of the process is building code that goes through defined steps. WDYT?

example/index.html
```diff
@@ -1,17 +1,35 @@
...
+        rel="noopener noreferrer"
+        >View project on GitHub</a
+      >
     </h2>
     <div id="app"></div>
     <div class="footer">
       Copyright &copy; 2016 - <span id="currentYear"></span>
-      <a href="https://github.com/P5-wrapper/react/graphs/contributors"
+      <a
+        href="https://github.com/jamesrweb"
```
yevdyko, line 23: 😄

.github/workflows/CD.yml
```diff
@@ -0,0 +1,52 @@
...
+          path: ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build the demo application
+        run: pnpm build:demo
```
yevdyko, line 47: 👍

yevdyko (changes requested):
This is a good idea! I left a few questions before we merge it to `master`.

.github/workflows/CI.yml
jamesrweb: I wanted to be a bit more explicit and perhaps it makes sense to even split the job into its constituent parts at some point but I don't know, I would leave it as it is, it's not a big deal and at least keeps the CI step a little more descriptive until the job is split up, WDYT?

example/index.html
jamesrweb: This is because of the NPM organisation. The names on GitHub and NPM are the same for the organisation, it just so happens that the name you install with is different (for now).

Both `npm install @P5-wrapper/react` and `npm install react-p5-wrapper` will be acceptable in the future since the current name will eventually just be an alias for the new one.

That's a whole other story though and will take time to implement and test properly. Perhaps this could be done as part of the version 4 release since it will be a major change version anyway. Let's see.

example/index.html
jamesrweb: See above comment. 🙇🏻‍♂️

.github/workflows/CI.yml
yevdyko: I don't mind if we leave it at that. It seems a bit odd, but if that's okay with you, I'm fine with it too :)

example/index.html
yevdyko: That makes sense now and thanks for the explanation! So, it looks like we're all set 🏎️

yevdyko (approved):
🚢

PR #177, opened 2022-08-05
Type updates and runtime error handling

### LINKED ISSUE #173: [Investigation] Remove `rooks` dependency
Currently we use `rooks` to expose to us an isomorphic use effect but it seems that almost nobody uses SSR with this library and thus, we could save space in the downloadable bundle by moving to a purely client side implementation.

I have some concerns with doing this such as integrations with libraries such as `remix`, `gatsby` and `next` but these need to be tested as part of the outcome, if any, of this issue.

Furthermore, as part of this issue, we should run some benchmarks to see how much, if any, impact is visible in the end bundle and on core use cases.

### LINKED ISSUE #176: Split the build CI job into smaller parts
At some point we would like to split the build CI job into smaller parts. The link to the discussion is [here](https://github.com/P5-wrapper/react/pull/174/files#r929623740)

jamesrweb:
Fixes #179
Fixes #176

Update project types and add a guard wrapper to log an appropriate error message if no sketch is provided at runtime.

## Proposed Changes

- Remove legacy types
- Add new types for guard handler
- Handle runtime exception case for a missing sketch prop.

This PR is more of a proposal, please review and provide feedback as usual, happy for design / implementation suggestions if any come to mind for discussion!

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -7,44 +7,22 @@ on:
 concurrency: cd-${{ github.ref }}
 
 jobs:
-  build-and-deploy:
+  build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-        name: Checkout the latest code for the current PR
+      - uses: P5-wrapper/setup-action@v1
         with:
           token: ${{ secrets.GH_TOKEN }}
 
-      - uses: actions/setup-node@v2
-        name: Setup node
-        with:
-          node-version: 16
-
-      - uses: pnpm/action-setup@v2.0.1
-        name: Install pnpm
-        id: pnpm-install
-        with:
-          version: 7
-          run_install: false
-
-      - name: Get pnpm store directory
-        id: pnpm-cache
-        run: |
-          echo "::set-output name=pnpm_cache_dir::$(pnpm store path)"
-
-      - uses: actions/cache@v3
-        name: Setup pnpm cache
-        with:
-          path: ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
-
-      - name: Install dependencies
-        run: pnpm install
-
       - name: Build the demo application
         run: pnpm build:demo
+
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: P5-wrapper/setup-action@v1
+        with:
+          token: ${{ secrets.GH_TOKEN }}
 
       - name: Deploy the demo application
         uses: JamesIves/github-pages-deploy-action@v4
--- a/config/eslint/eslint.config.js
+++ b/config/eslint/eslint.config.js
@@ -6,7 +6,6 @@ module.exports = {
   extends: [
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
-    "plugin:prettier/recommended",
     "plugin:react-hooks/recommended",
     "plugin:react/recommended"
   ],
@@ -18,7 +17,7 @@ module.exports = {
     ecmaVersion: 12,
     sourceType: "module"
   },
-  plugins: ["@typescript-eslint", "prettier", "react", "simple-import-sort"],
+  plugins: ["@typescript-eslint", "react", "simple-import-sort"],
   overrides: [
     {
       files: [".js", ".jsx", ".ts", ".tsx"]
@@ -28,10 +27,7 @@ module.exports = {
     "@typescript-eslint/no-explicit-any": "off",
     "linebreak-style": ["error", "unix"],
     "simple-import-sort/exports": "error",
-    "simple-import-sort/imports": "error",
-    indent: ["error", 2],
-    quotes: ["error", "double"],
-    semi: ["error", "always"]
+    "simple-import-sort/imports": "error"
   },
   settings: {
     react: {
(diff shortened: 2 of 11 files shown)


config/eslint/eslint.config.js
```diff
@@ -6,7 +6,6 @@ module.exports = {
   extends: [
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
-    "plugin:prettier/recommended",
```
jamesrweb, line 9: We don't need prettier in eslint since prettier runs and validates itself.

src/index.tsx
```diff
@@ -1,13 +1,15 @@
...
-import React, { createRef, memo, MutableRefObject, useRef } from "react";
+import React from "react";
 import { useIsomorphicEffect } from "rooks";
 
 type Wrapper = HTMLDivElement;
 type WithChildren<T = unknown> = T & { children?: React.ReactNode };
 type InputProps<Props extends SketchProps = SketchProps> = Props & {
   sketch?: Sketch<Props>;
 };
+type P5WrapperPropsWithSketch<Props extends SketchProps = SketchProps> =
```
jamesrweb, line 10: Guard case to ensure a sketch is provided.

src/index.tsx
```diff
@@ -18,12 +20,11 @@ export type P5CanvasInstance<Props extends SketchProps = SketchProps> = p5 & {
   updateWithProps?: (props: Props) => void;
 };
 
-// @TODO: remove in next major version, keep for compatibility reasons for now.
-export type P5Instance<Props extends SketchProps = SketchProps> =
```
jamesrweb, line 22: Remove legacy type in preparation for the next major release.

src/index.tsx
```diff
@@ -33,23 +34,32 @@ function createCanvasInstance<Props extends SketchProps = SketchProps>(
...
 function removeCanvasInstance<Props extends SketchProps = SketchProps>(
-  canvasInstanceRef: MutableRefObject<P5CanvasInstance<Props> | undefined>
+  canvasInstanceRef: React.MutableRefObject<P5CanvasInstance<Props> | undefined>
 ) {
   canvasInstanceRef.current?.remove();
   canvasInstanceRef.current = undefined;
 }
 
-function ReactP5WrapperComponent<Props extends SketchProps = SketchProps>({
+function ReactP5WrapperComponentGuard<Props extends SketchProps = SketchProps>({
```
jamesrweb, line 42: Add a guard to catch runtime validations of input data, in this case the sketch is what we are looking to validate.

src/index.tsx
```diff
@@ -33,23 +34,32 @@ function createCanvasInstance<Props extends SketchProps = SketchProps>(
...
   sketch,
-  children,
   ...props
 }: P5WrapperProps<Props>) {
   if (sketch === undefined) {
-    throw new Error("[ReactP5Wrapper] The `sketch` prop is required.");
+    console.error("[ReactP5Wrapper] The `sketch` prop is required.");
```
jamesrweb, line 47: Error would break a users client scripts potentially, error logging and empty render node seems better for DX / UX. WDYT?

src/index.tsx
```diff
@@ -33,23 +34,32 @@ function createCanvasInstance<Props extends SketchProps = SketchProps>(
...
+
+    return null;
   }
 
-  const wrapperRef = createRef<Wrapper>();
-  const canvasInstanceRef = useRef<P5CanvasInstance<Props>>();
+  return <ReactP5WrapperComponent sketch={sketch} {...props} />;
```
jamesrweb, line 52: Sketch is now validated and we can render freely.

tests/rendering.test.tsx
```diff
@@ -67,17 +67,13 @@ describe("Rendering", () => {
     expect(StaticComponent).toBe('<div class="react-p5-wrapper"></div>');
   });
 
-  it("[General] should throw an error when the `sketch` prop is not provided", () => {
-    expect(() => render(<ReactP5Wrapper />)).toThrow(/sketch/);
-  });
-
-  xit("[General] Should not render anything when the `sketch` prop is not provided", () => {
+  it("[General] Should not render anything when the `sketch` prop is not provided", () => {
```
jamesrweb, line 69: We don't need to skip these tests anymore now that the implementation is done.

tests/updates.test.tsx
```diff
@@ -1,10 +1,10 @@
 import { render } from "@testing-library/react";
 import React from "react";
 
-import { P5Instance, ReactP5Wrapper, Sketch } from "../src";
+import { P5CanvasInstance, ReactP5Wrapper, Sketch } from "../src";
```
jamesrweb, line 4: Swap legacy type for preferred alternative.

jamesrweb:
@yevdyko ready for review, no more changes will be made! When you get some time, take a look and let me know what you think 👀.

.github/workflows/CD.yml
```diff
@@ -7,45 +7,23 @@ on:
 concurrency: cd-${{ github.ref }}
 
 jobs:
-  build-and-deploy:
+  build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-        name: Checkout the latest code for the current PR
+      - uses: P5-wrapper/setup-action@v1
```
yevdyko, line 13: 🍦

src/index.tsx
yevdyko: I think this can be a good idea because it's the fault of the developers, not the users, so even if this error is handled correctly on the client, it doesn't tell the user anything

yevdyko (approved):
Looks good to me @jamesrweb! Let's release these changes 🚢

PR #178, opened 2022-08-13
Audit dependencies

### LINKED ISSUE #173: [Investigation] Remove `rooks` dependency

jamesrweb:
Fixes #173

## Proposed Changes

- Remove `rooks` dependency and use regular `useEffect` again
- Remove unused dev dependencies
- Sort keys in package.json blocks
- Remove unused test streams helper
- Change TS output location to avoid duplicated `index.d.ts` on build
- Update and apply a security audit to remaining dependencies

N / A

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,7 +1,6 @@
 import diff from "microdiff";
 import p5 from "p5";
 import React from "react";
-import { useIsomorphicEffect } from "rooks";
 
 type Wrapper = HTMLDivElement;
 type WithChildren<T = unknown> = T & { children?: React.ReactNode };
@@ -61,7 +60,7 @@ function ReactP5WrapperComponent<Props extends SketchProps = SketchProps>({
   const wrapperRef = React.createRef<Wrapper>();
   const canvasInstanceRef = React.useRef<P5CanvasInstance<Props>>();
 
-  useIsomorphicEffect(() => {
+  React.useEffect(() => {
     if (wrapperRef.current === null) {
       return;
     }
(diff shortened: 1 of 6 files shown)


src/index.tsx
```diff
@@ -61,7 +60,7 @@ function ReactP5WrapperComponent<Props extends SketchProps = SketchProps>({
   const wrapperRef = React.createRef<Wrapper>();
   const canvasInstanceRef = React.useRef<P5CanvasInstance<Props>>();
 
-  useIsomorphicEffect(() => {
+  React.useEffect(() => {
```
yevdyko, line 63: How about making the use of `rooks` optional? 

By default we use `UseEffect`, and by passing the parameter you can use `useIsomorphicEffect`.

WDYT? Or bundle size more important from your point of view?

jamesrweb: Actually the server tests work as expected with the regular use effect at the moment and so an isomorphic effect isn't required. This is because of the changes we made to move to refs meaning everything can be done client side and all we render server side is the wrapper div element.

jamesrweb: You could, if you really wanted to, test with a nextjs app to see for yourself how it would work as part of your testing of the PR but it's up to you ☺️

yevdyko: I don't want to go crazy like that, so I think we can merge it as it is for now.

yevdyko (approved):
thanks for this updates @jamesrweb 🚢 🚢 🚢

yevdyko:
@jamesrweb I was wondering if this PR can go directly to the main?

jamesrweb:
@yevdyko its based on the other branch so sadly not. Merging now since you approved. 😊

PR #179, opened 2022-08-13
Split CI Jobs

### LINKED ISSUE #176: Split the build CI job into smaller parts

jamesrweb:
Fixes #176

## Proposed Changes

- Split composite job into individual jobs
- Utilise [the new setup job](https://github.com/P5-wrapper/setup-action) which each other job depends upon

N / A


.github/workflows/CD.yml
```diff
@@ -7,20 +7,20 @@ on:
 concurrency: cd-${{ github.ref }}
 
 jobs:
-  build-and-deploy:
+  setup:
```
yevdyko: We use the same `setup` in CI and CD workflows. What about using action composition to reduce duplication? Is this possible here?

https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/

jamesrweb: The current setup I've made doesn't even work, it seems not likely to either. Once I get things working somehow, then I will take a look at that, thanks for the tip! Seems interesting😇

jamesrweb: Created a shared job for both places 🚀!

jamesrweb:
@yevdyko working! 🚀

Please review the two open PRs as soon as you get time. 🎁

jamesrweb:
@yevdyko I moved the base because the other PR is bugging out waiting for the build task that is required in settings now and so it needs the new flows to correct itself 👍🏻

jamesrweb:
The PR automerged due to repo settings 🤷🏻‍♂️... didn't expect that actually... but it's okay since the target was the other PR anyway 😄

PR #180, opened 2022-08-27
Fix for demo deployments in the CD workflow

jamesrweb:
## Proposed Changes

- Merge jobs to resolve deployment issue 🤞🏻
- Update dependencies

Jobs run independently and so although the build was successful, the deployment job had no idea how to actually deploy since the `dist/demo` directory didn't exist as far as it was concerned.


yevdyko (approved):
Thanks for the fix @jamesrweb! ⛵

PR #181, opened 2022-08-28
Auto deploy to NPM in CD workflow

jamesrweb:
## Proposed Changes

- Automatically deploy to NPM on each master merge

Please read the [documentation](https://github.com/marketplace/actions/npm-publish) and suggest alternatives if you know of any better options or can find any. To me this seemed to be the simplest and offers a nice benefit (so it says) that if the version didn't change, it won't deploy at all which means we can control if it deploys or not purely by controlling the version within the `package.json` file.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -7,7 +7,7 @@ on:
 concurrency: cd-${{ github.ref }}
 
 jobs:
-  deploy:
+  gh-pages:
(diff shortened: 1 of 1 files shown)


yevdyko (approved):
I think this might be a good way to automate the publishing process. If everything works as expected, you don't have to do it manually. So let's give it a try 🎸

PR #184, opened 2022-09-07
Automate project dependency auditing and updating

jamesrweb:
## Proposed Changes

- Automate project dependency auditing
- Automate project dependency updating
- Run the automation for dependency management on a weekly basis


yevdyko (commented):
I'm wondering if we can achieve the same thing by [configuring a dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)? Is there any advantage to using your own cron job script?

jamesrweb:
> I'm wondering if we can achieve the same thing by [configuring a dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file)? Is there any advantage to using your own cron job script?

I'm honestly not sure, I guess the main benefit is just full customisation and to continue using `pnpm` with our setup script. Perhaps Dependabot is a better direction in the sense that it's a GH product though... WDYT?

yevdyko:
> I'm honestly not sure, I guess the main benefit is just full customisation and to continue using `pnpm` with our setup script. Perhaps Dependabot is a better direction in the sense that it's a GH product though... WDYT?

Okay, I've read about it [in more detail](https://docs.renovatebot.com/javascript/). At the moment `dependabot` doesn't support `pnpm`, so we won't be able to use it, but we can try [renovatebot](https://docs.renovatebot.com/javascript/) and for now let's still try your script. How about this?

jamesrweb:
> > I'm honestly not sure, I guess the main benefit is just full customisation and to continue using `pnpm` with our setup script. Perhaps Dependabot is a better direction in the sense that it's a GH product though... WDYT?
> 
> 
> 
> Okay, I've read about it [in more detail](https://docs.renovatebot.com/javascript/). At the moment `dependabot` doesn't support `pnpm`, so we won't be able to use it, but we can try [renovatebot](https://docs.renovatebot.com/javascript/) and for now let's still try your script. How about this?

Sure, we can try this script first and then look into renovate later then as you suggested 👍🏻

yevdyko (approved):

PR #197, opened 2022-11-02
Reset sourcemap style

jamesrweb:
## Proposed Changes

- Reset sourcemap style to keep consistency with previous releases

This was altered due to an issue in #194 but should be fine now.


yevdyko (approved):
👍

Issue #207, opened 2022-11-24
Known typescript sub-typing issues


jamesrweb:
## Expected Behavior

Typescript gets smart.

## Actual Behavior

Typescript is not smart.

## Steps to Reproduce the Problem

1. View `src/components/ReactP5WrapperGuard.tsx`.
2. View `src/components/ReactP5WrapperWithSketch.tsx`.
3. Notice the `as unknown as T` shenanigans.

The `as unknown as T` casts are begrudgingly required due to a known limitation of the TypeScript compiler as demonstrated in the following issues for example:

-   [typescript/#35858](https://github.com/microsoft/TypeScript/issues/35858)
-   [typescript/#37670](https://github.com/microsoft/TypeScript/issues/37670)

 As Ryan Cavanaugh points out:

> "TS isn't capable of the higher-order reasoning needed to understand that `Exclude<T, k> & { [k]: T[k] }` is equivalent to `T`."
>
> **Note**: this is the same issue as we face here just with `Omit` instead of `Exclude`.

Potentially this would have been resolved by this PR but the author closed it due to lack of time:

-   [typescript/#42382](https://github.com/microsoft/TypeScript/pull/42382)

Perhaps someone interested in the topic and who has time to work on it would be willing to take a look into the issue.

Either way, until a resolution is merged into the TypeScript compiler that addresses this issue, we need to use this workaround.  We could also remove this if we manage find a reasonable, more fitting workaround of some sort to avoid casting in the first place. If a workaround / change of implementation comes to mind, please raise an issue on the repository or feel free to open a PR!

## Specifications

Package Version: 4.0.0 and above

PR #229, opened 2023-02-20
Allow Shallow Comparison based on deepCompare prop

### LINKED ISSUE #227: Problem with updateWithProps deeply comparing object.
## Expected Behavior
passing an object prop to `<ReactP5Wrapper>` triggers shallow comparison and re-render occurs only when address reference of object changes

## Actual Behavior
 passing as prop a large object (Tone.js synth) causes huge prop comparison/ re-rendering time (over 2s), makes application unusable. 

## Specifications:
I am using a p5 and Tone.js to create a music related app. Currently working on connecting an 'instrument' component (a p5js interactive sketch that allows users to create a note sequence) to a sound source (a synthesizer component using Tone.js). I want to pass the Tone.js synth object (that is controlled by the synth component) to the 'instrument' component ,so that it can handle triggering the synth according to the note sequence inputted by the user. 

However, when passing a tone.js Polysynth instance as props to `<P5ReactWrapper/>` the deep comparison that occurs causes huge delay times in re-rendering, given that Polysynth is a fairly large Object. This is particularly problematic as in my case I don't care about whether the difference lies within the synth's **parameters** - I only want to it to check for changes in the synth prop's **adress** (i.e. wether the user want's to connect the instrument to a different synth instance.) 

Regular React components handle this case perfectly, but the way `<ReactP5Wrapper>` handles prop updates (src/utils/propsAreEqual.ts) causes many performance issues in my use case. 

Is there a way I could bypass the deep comparison for specific `<ReactP5Wrapper>` instances or maybe would it be possible to implement that possibility with a PR?

Thank you !

Package Version:
4.1.0

manolara:
Fixes #227

## Proposed Changes

- Adds prop to `<ReactP5Wrapper>` called `deepCompare`.
- Changes the Logic of `propsAreEqual`. If `deepCompare` is not passed or if it is set  to `true` behaviour is same as before. If `deepCompare` is set to `false`, `propsAreEqual` returns `undefined`, hence memo(ReactP5WrapperGuard, propsAreEqual) only has one parameter and handles prop comparison in the standard react way (shallow comparison).
- Installed Tone.js and edited `Box.js` example and `App.jsx` to replicate my Tone.js synth problem and to demonstrate how the `deepCompare` prop solves the issue. If we reach a point of being happy and can proceed to merge, all these changes will be of course removed.

## Additional Notes
In order to get this to work, I had to add an `as any` in the memo() function in `ReactP5Wrapper.tsx` which is by no means a best practise. Let me know if you have any good work-arounds or if this is acceptable/unacceptable.


example/sketches/box.js
```diff
@@ -1,12 +1,31 @@
+import * as Tone from "tone";
```
jamesrweb: Don't update existing examples, instead create a new one.

package.json
```diff
@@ -74,7 +74,8 @@
   },
   "dependencies": {
     "microdiff": "^1.3.1",
-    "p5": "^1.5.0"
+    "p5": "^1.5.0",
+    "tone": "^14.7.77"
```
jamesrweb: tone.js is not a dependency of the library, it should not be here but instead in dev-dependencies since you only want it for the example.

src/components/ReactP5Wrapper.tsx
```diff
@@ -3,4 +3,4 @@ import { memo } from "react";
 import { propsAreEqual } from "../utils/propsAreEqual";
 import { ReactP5WrapperGuard } from "./ReactP5WrapperGuard";
 
-export const ReactP5Wrapper = memo(ReactP5WrapperGuard, propsAreEqual);
+export const ReactP5Wrapper = memo(ReactP5WrapperGuard, propsAreEqual as any);
```
jamesrweb: We never use `as` without an unavoidable issue such as the one outlined in #207.

src/utils/propsAreEqual.ts
```diff
@@ -7,7 +7,13 @@ export function propsAreEqual<Props extends SketchProps = SketchProps>(
...
-  const differences = diff(previous, next);
+  // set deepCompare to true if not specified by props
+  const deepCompare = next.deepCompare === undefined ? true : next.deepCompare;
 
-  return differences.length === 0;
+  if (deepCompare) {
+    const differences = diff(previous, next);
+    return differences.length === 0;
+  } else {
+    return undefined;
```
jamesrweb: There needs to still be diffing of some sort done here, otherwise every frame will trigger a re-render. I am also unsure if it is wise to pass a prop to control this behaviour, I think it should be done as part of a provider of some sort for example which allows hooking into the config. Perhaps a better option is to simply check the size of the props being passed in and above a certain limit, deciding which route is better (deep or shallow comparison). I am still unsure that this overall is a good idea though because in the past shallow comparisons caused a lot of re-rendering issues, we need to be really careful with this.

jamesrweb (changes requested):
Please also ensure that new test cases are added for any of the changes to the library. 🙏🏻✅

src/utils/propsAreEqual.ts
manolara: Ok that mostly makes sense, but could you please elaborate on why passing a prop to control that behaviour would not be a good idea?

jamesrweb: Because the props are meant to be passed to the sketch, not control the generation of the sketches render, does that make sense?

manolara:
I kinda messed up with git big time 🤦‍♂️ and i've squashed more commits than I should've. My fork has become way more convoluted than it should be and reverting my mistakes seems way more time consuming than necessary. I'm going to make a new fork and a new PR to keep things neat. I've considered your comments in the new PR.

src/utils/propsAreEqual.ts
manolara: Yes, it does makes sense! I completely understand your concerns regarding both the role that`deepCompare` prop plays as well as the performance issues that my changes may impose. Nonetheless, I'm curious to see what you think about how my additions work in the  example I implemented (4th sketch in `app.tsx`). With the current changes in `propsAreEqual`, the sketch re-renders only when the props change so that seems ok to me. I also removed the `as any` considering that `propsAreEqual` should now only return a bool. However, If you're still not convinced that going down this path is a good idea, I'm more than happy to explore different options. 👍

example/app.jsx
```diff
@@ -53,10 +55,27 @@ function App() {
       </Fragment>
     );
   }
+  const synth1 = useMemo(
```
jamesrweb, line 59: This could all be inside the tone example, right?

example/app.jsx
```diff
@@ -65,8 +84,15 @@ function App() {
         step="1"
         onChange={onRotationChange}
       />
+
       <button onClick={onChangeSketch}>Change Sketch</button>
       <button onClick={onMountStateChange}>Unmount</button>
+      {state.sketch === synth.sketch ? (
+        <>
```
jamesrweb: And this could be inlined within the tone example also, right?

example/app.jsx
```diff
@@ -53,10 +55,27 @@ function App() {
...
   return (
     <Fragment>
-      <ReactP5Wrapper sketch={state.sketch} rotation={state.rotation} />
+      <ReactP5Wrapper
+        sketch={state.sketch}
+        rotation={state.rotation}
+        deepCompare={false}
+        synth={currentSynth}
+      />
       <input
```
jamesrweb, line 80: I would add `state.sketch !== synth.sketch` for this input since we don't use the rotation input only for that example, right?

src/utils/propsAreEqual.ts
```diff
@@ -7,7 +7,25 @@ export function propsAreEqual<Props extends SketchProps = SketchProps>(
   previous: P5WrapperProps<Props>,
   next: P5WrapperProps<Props>
 ) {
-  const differences = diff(previous, next);
+  // set deepCompare to true if not specified by props
+  const deepCompare = next.deepCompare === undefined ? true : next.deepCompare;
```
jamesrweb:
```suggestion
  const deepCompare = next.deepCompare ?? true;
```

jamesrweb: We can just use the null coalescence operator for this use case to get the same effect, right?

src/contracts/InputProps.ts
```diff
@@ -3,4 +3,5 @@ import { type SketchProps } from "./SketchProps";
 
 export type InputProps<Props extends SketchProps = SketchProps> = Props & {
   sketch?: Sketch<Props>;
+  deepCompare?: boolean;
```
jamesrweb:
```suggestion
  useDeepObjectComparisonChecksWhenDiffingPropsForNextRender?: boolean;
```

jamesrweb: I don't know if my suggestion is even the best name, I just want it to be clear what that prop actually does. If you have a cleaner or better idea, please do suggest it instead! Oh and we also need to document this prop either way in the README file, with an example, just like the others are.

src/utils/propsAreEqual.ts
```diff
@@ -7,7 +7,25 @@ export function propsAreEqual<Props extends SketchProps = SketchProps>(
...
+        return false;
+      }
+    }
+    for (const key in next) {
+      if (!(key in previous) && next[key] !== undefined) {
+        return false;
+      }
+    }
+    return true;
+  }
```
jamesrweb: I would move these to their own function definition above of the `propsAreEqual` definition. Something like:

```ts
function shallowEqual<T>(previous: T, next: T) {
  const previousKeys = Object.keys(previous);
  const nextKeys = Object.keys(next);

  if (previousKeys.length !== nextKeys.length) {
    return false;
  }

  for (const key of previousKeys) {
     if (previous[key] !== next[key]) {
       return false;
     }
   }

  return true;
}

function deepEqual<T>(previous: T, next: T) {
  return diff(previous, next).length === 0;
}

export function propsAreEqual<Props extends SketchProps = SketchProps>(
   previous: P5WrapperProps<Props>,
   next: P5WrapperProps<Props>
) {
  const deepCompare = next.deepCompare ?? true;

  return deepCompare ? deepEqual(previous, next) : shallowEqual(previous, next);
}
```

Something like that but you can adapt it as you like. I just think the abstraction makes things a bit cleaner. Also note that I wrote this here and didn't actually test this code so please check and test and update if necessary.

jamesrweb (changes requested):
The PR looks much better now, great work! I have some final suggestions, once you get chance to take a look and alter as you see fit, I can re-review and release 👍🏻!

src/utils/propsAreEqual.ts
jamesrweb: Yeah, I see your point also, for now we can try it your way and see how it goes 🚀

jamesrweb:
You can ignore the `CI/renovate-bot` failure, it's a fork issue and not something you introduced 😄!

src/utils/propsAreEqual.ts
manolara: Hey, so sorry for being inactive on the matter for a while. Interestingly, I managed to find a work-around. Instead of passing down the synth object - i pass down a function, that does absolutely nothing, but return the synth object.. This seems to have resolved the issue and by using useCallback I can change the function when the synth address changes hence mimick a shallow comparison. Up to now that seems  to work fine and has not introduced any problems . It is however a not very elegant solution of course, neither very idiomatic, so if you want I can keep working on this PR to setup a cleaner way to bypass deep comparisons. Let me know what you think

jamesrweb: Yeah, keep exploring this PR if you're up for it, I'm curious what you come up with. I'm also curious to see your current implementation described above if you have a link. Indeed it is a solution also.

example/app.jsx
manolara: Yes, it could be - however adding the synth instances inside the synth example would defy the purpose of the example altogether. With this added synth.js example I aimed to showcase how P5Wrapper handles large objects as props and the effect that the new added deepCompare prop will have on that. I chose the tone.js synths as the large objects for this demo  as they are long enough to cause significant delays when deep diffing. That's why it's necessary that the synth instances live on `App.jsx`, so that they can actually be passed down to the synth sketch through props and showcase how the deepCompare prop affects the performance. Otherwise the deepCompare prop wouldn't be put to action.
hope that makes sense

example/app.jsx
manolara: I intentionally left it. When having large objects passed as props, interacting with other "normal" props such as rotation, will still cause the huge delays. Changing the rotation value, in this case, will still cause a complete prop diffing hence they synth object is deep compared and the app will completely lag for 1-3 seconds even when interacting with props unrelated to the synth such as the rotation slider.

example/app.jsx
manolara: I intentionally left it. When having large objects passed as props, interacting with other "normal" props such as rotation, will still cause the huge delays. Changing the rotation value, in this case, will still cause a complete prop diffing hence they synth object is deep compared and the app will completely lag for 1-3 seconds even though the user may have interacted with props unrelated to the synth like the rotation slider.

src/utils/propsAreEqual.ts
manolara: the implementation of that solution is truly very simple, its just that:

image

example/app.jsx
manolara: same goes for this

src/utils/propsAreEqual.ts
manolara: Added a new PR and addressed your comments, let me know what you think :)

manolara: Added a new PR and addressed your comments, let me know what you think :)

jamesrweb (approved):
Approving and merging to a sandbox branch for testing. Will update the issue #229 when testing is complete and merged to master for release.

jamesrweb:
@manolara actually, you need to resolve the conflicts first, can you sync and resolve so I can merge this?

manolara:
> @manolara actually, you need to resolve the conflicts first, can you sync and resolve so I can merge this?

Seems like the pnpm-lock.yaml got some format changes and i couldn't find a way to reformat it. Dont know if thats a problem.

PR #248, opened 2023-04-30
Add the ability for fallback UIs to be rendered if the sketch is ever undefined

jamesrweb:
This PR allows a fallback UI to be rendered if the sketch is ever undefined rather than purely returning `null` and thus rendering nothing although this will still be the default behaviour if no fallback is provided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,80 @@ Of course, you can also use any other css-in-js library or by just using simple
 css to achieve almost anything you can imagine just by using the wrapper class
 as your root selector.
 
+## Fallback UIs
+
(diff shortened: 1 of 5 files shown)


yevdyko (commented):
The code looks good, but I was wondering if this option should be described in README. @jamesrweb WDYT?

jamesrweb:
> The code looks good, but I was wondering if this option should be described in README. @jamesrweb WDYT?

True. That's a good point, I'll update the docs and push later 🙏🏻

yevdyko (approved):
@jamesrweb Thanks for the feature and documentation!

PR #251, opened 2023-05-02
Update dependencies

jamesrweb:


package.json
```diff
@@ -85,27 +85,29 @@
...
     "typescript": "^5.0.4",
-    "vite": "^4.2.2",
+    "vite": "^4.3.4",
     "vitest": "^0.30.1"
   },
   "dependencies": {
     "microdiff": "^1.3.2",
-    "p5": "^1.6.0"
+    "p5": "^1.6.0",
+    "react": "^18.2.0",
```
yevdyko: Do we need to have React as a dependency in the package?

jamesrweb: I think `PNPM` added these itself for some reason during the `upgrade` command execution cycle or perhaps it was `ncu -u` which did it. Not sure what happened honestly. Either way, something automated, somewhere, made the change but I removed them from dependencies now.

jamesrweb: Good catch 🚀

jamesrweb: Need to resolve conflicts when I get onto my laptop, made the change via the mobile app and forgot to update the lock file which is why the CI is on blocked status 😅

yevdyko (approved):
@jamesrweb Thanks!

🚢

PR #280, opened 2023-10-22
Use dependabot instead of renovate

yevdyko:
This PR adds dependabot configuration instead of renovate, since Github started supporting `pnpm` and it's a bit more convenient to use the built-in dependency management tool.

Hope you find this useful @jamesrweb 

Reference: [Dependabot version updates now supports pnpm](https://github.blog/changelog/2023-06-12-dependabot-version-updates-now-supports-pnpm/)

## Proposed Changes

- Add dependabot config
- Remove renovate config
- Update workflow config for formatting PRs from bot


jamesrweb:
Can you also make a PR for the next and setup-action repos also? I want to keep all repos in sync and so we should try to update them all at the same time where possible then.

Otherwise, looks good to me! 😊🚀

PR #298, opened 2023-12-06
Update dependencies for patch release

jamesrweb:
Updates all dependencies except `vitest` since `vitest` breaks when using a version greater than `1.x.x`. This needs investigated but is probably something to do with the issue from the `vite` / `rollup` infrastructure, just like the issues with `vite` versions less than `vite@5.0.4` in the `5.x.x` range.


yevdyko (approved):
🚢

PR #494, opened 2025-02-05
change typescript module resolution to resolve typescript resolution …

### LINKED ISSUE #493: type ReactP5WrapperGuard isn't properly resolved when building

Qanpi:
…errors on build (#493)

Fixes #493

## Proposed Changes

- change module resolution to 'node', bundler fails to properly resolve type imports for whatever reason
- change default function exports to const and default exports (again, seems to be necessary probably because of some unique bundler issue)

### DIFF
--- a/config/typescript/tsconfig.json
+++ b/config/typescript/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "target": "ESNext",
-    "composite": true,
     "useDefineForClassFields": true,
     "types": ["vite/client"],
     "lib": ["DOM", "DOM.Iterable", "ESNext"],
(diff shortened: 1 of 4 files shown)


jamesrweb (approved):

yevdyko:
FYI: While I was testing, I discovered the following. It seems that the changes in the default function export are not related to the failed build, but they can be updated too.

yevdyko:
FYI @jamesrweb @Qanpi I played around a bit with `tsconfig` and the only thing that matters is defining the `composite` value as `false`, the module resolution type has no effect on build errors and works fine in a more modern option like `bundler`.

PR #510, opened 2025-06-01
Fix CI issues

jamesrweb:
## Proposed Changes

- Downgrade `p5` from version 2.x.x to version 1.x.x
  - The changes there need a new major version to be introduced
  - I am also considering adding an option to use `q5` in future and support both in this library.
  - This was the core issue with the CI as tests couldn't pass with the version 2.x.x changes.
- Remove `ts-ignores` in the `eslint` config now that the types are finally exposed by the `react` team.
- Update `pnpm` dependencies and allow builds for local systems of chosen dependencies to allow CI and local to be in sync irregardless of OS.

N / A


yevdyko (approved):

yevdyko:
It's definitely a good idea to add `q5` as a new option 💪

PR #517, opened 2025-07-18
Github specific workflow refactors and documentations

jamesrweb:
## Proposed Changes

- Update configurations in the `.github` directory
- Adapt the Dependabot workflow
- Try to make CI / CD be more declarative

I would like to discuss the idea of enforcing [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) so that we can also auto-generate a changelog but what do you think of this @yevdyko? I think this would help to add more context for releases but it does mean the overhead of enforcing the commit style to achieve that end. That's the main trade-off but I think it could be worthwhile, I'm interested in your feedback on that.

Also, what is your plan for the `workflows` repository and the `p5` 2.x.x migration we discussed? As an update from my side, the `q5` flows are almost finished and I just need to adapt the tests and change a couple of the types now before making the PR for that topic.


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,13 +1,55 @@
...
+## Issue Type
+
+<!-- Please check one or more that apply by replacing [ ] with [x] -->
+
+- [ ] 🐛 Bug Report
+- [ ] 💡 Feature Request
+- [ ] 📝 Documentation Issue
+- [ ] 🤔 Question
+- [ ] 🧹 Chore
+- [ ] ❓Other
```
yevdyko, line 15:
```suggestion
- [ ] ❓ Other
```

yevdyko (approved):
Thanks for the improvements — everything looks good! Regarding conventional commits, I’m personally not a big fan of them, especially for small projects with few contributors. That said, I don’t mind if you’d like to adopt this approach. The only thing I’d ask: could we avoid using emojis in commit messages? It feels a bit too much for my taste 🙂

Since you've updated all the workflow configurations here, I can extract the shared ones to the `workflows` repository after this PR is merged.

Could you remind me of the details about the p5 2.x.x migration? This topic slipped my mind

jamesrweb:
We need to upgrade p5 to v2.x.x and adapt implementation & the tests. I guess it makes sense to go into v5 before it's properly released since it's a breaking change. Or we release v5 as is and make v6 for the p5 upgrade but I don't think it makes sense to do 2 major upgrades in the space of a few weeks / months. What do you think?

PR #523, opened 2025-08-18
Ensure the project readme is deployed to NPM also

jamesrweb:
## Related Issue

N / A

## PR Type

- [x] 🐛 Bug Fix
- [ ] ✨ New Feature
- [ ] 🔨 Code Refactor
- [ ] 📝 Documentation Update
- [ ] 🧪 Test Update
- [ ] 🔧 Build/CI Update
- [ ] 🧹 Chore
- [ ] ⏪ Revert

## Description

In the last deployment for `v5.0.0-rc.4`, the project `README.md` was not deployed since it is now in a different root directory.

## Proposed Changes

- Copy the readme to the root pre-deploy
- Remove it from the root post-deploy
- Add it to the list of `files` in the `package.json` to make it mandatory

## How Has This Been Tested?

- [ ] Unit Tests
- [ ] Integration Tests
- [x] Manual Testing (please describe)

Tested by running `pnpm publish --tag next --dry-run --no-git-checks`.

## Screenshots/Recordings

N / A

## Breaking Changes

- [ ] Yes (please describe)
- [x] No

## Checklist

- [x] My code follows the code style of this project
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests passed
- [x] My changes generate no new warnings

## Additional Notes

N / A

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "homepage": "https://github.com/P5-wrapper/react",
   "license": "MIT",
   "files": [
+    "README.md",
     "dist/component/*"
   ],
   "types": "./dist/component/main.d.ts",
@@ -28,6 +29,8 @@
     "integrate": "pnpm format:check && pnpm lint && pnpm test && pnpm build",
     "lint": "eslint --config config/eslint/eslint.config.ts",
     "lint:fix": "pnpm lint --fix",
+    "prepublishOnly": "cp .github/README.md ./README.md",
+    "postpublish": "rm ./README.md",
     "prettier": "prettier --config config/prettier/prettier.json --ignore-path .gitignore",
     "test": "vitest run --silent --config config/vite/vite.config.ts",
     "test:coverage": "vitest run --silent --coverage --config config/vite/vite.config.ts",


package.json
```diff
@@ -28,6 +29,8 @@
     "integrate": "pnpm format:check && pnpm lint && pnpm test && pnpm build",
     "lint": "eslint --config config/eslint/eslint.config.ts",
     "lint:fix": "pnpm lint --fix",
+    "prepublishOnly": "cp .github/README.md ./README.md",
```
yevdyko, line 32: Curious, why such a complication and why not put it in the root of the repository as usual? I think the files inside the ".github" directory only relate to the Github configuration and README is needed for documentation regardless of the provider. What don't I know?

jamesrweb: Same discussion we had about the configs, I just prefer as much to not be in the root directory as possible and `.github` is one of three places it can't be kept, so it's grouped with other related documents. Besides, the pre and post setup is standard to NPM itself and so works well for the use case anyway since it's only an issue for NPM deployments anyway.

yevdyko: I think this bug highlights that the current solution might be more of a workaround than an improvement and now we’re considering adding an extra hack on top.

According to the GitHub style guide, the README “should be located in the top-level directory for your product or library’s actual codebase”:

https://google.github.io/styleguide/docguide/READMEs.html#what-to-put-in-your-readme

If we move it to the root, contributors and visitors can immediately understand the project without digging into subfolders.

We could technically add a copy command in a prepublishOnly script, but that only runs right before publishing. It wouldn’t help with local development or repo clarity, and it adds unnecessary complexity. Keeping the README in the root is simpler, avoids extra steps and follows convention. Why not keep it simple? 😀

jamesrweb: The README is read out of the GitHub directory for the repo and temporarily copied for the NPM deployment. I honestly don't see an issue 😅. I do see your point that the two extra scripts are not _needed_ but I don't think it's an issue to have them and it keeps the root cleaner to me. In saying that, if you insist on it being in the root for conventions sake... maybe. I'm still not convinced but if you insist I suppose I can live with that argument but I still prefer it tidied into the GitHub directory personally since it's a supported location and it's only NPM that needs the deploy scripts which have 0 overhead. Let me know what your opinion is and we can see what to do 🤷🏻‍♂️￼.

jamesrweb:
Merged for now. We can discuss moving the README to the root again as a follow up, want to make sure people can actually see the docs on NPM for now though.

